PR #2, opened 2020-07-25
Attach methods to Tensor

wangkuiyi:


01-backward.go
```diff
@@ -6,49 +6,53 @@ package main
 // #include "cgotorch.h"
 import "C"
 
-func RandN(rows, cols int, require_grad bool) C.Tensor {
+type Tensor struct {
```
wangkuiyi, line 9: We can write `type Tensor C.Tensor`; however, this kind of type definition cannot have attached methods.

PR #13, opened 2020-07-27
Travis CI in both Linux and macOS

wangkuiyi:


wangkuiyi:
This configuration fails with the Travis CI bionic VM image because the Go version is 1.11, which is too old to understand `go env -w`.

The Go version in the osx image is new enough -- 1.13, so it passes.

PR #14, opened 2020-07-28
expose SGD optimizer

QiJune:
This PR exposes the optimizer interface of libtorch to Go.  After run `go run 01-backward.go`, we get the following log:

```
Parameter value:
-0.2994
-0.2610
-1.1278
 0.1280
[ CPUFloatType{4,1} ]
Gradient value:
-1.6246
-1.1954
 0.7638
-0.4145
[ CPUFloatType{4,1} ]
Parameter value after updating:
-0.1369
-0.1414
-1.2042
 0.1695
[ CPUFloatType{4,1} ]
```

The learning rate of SGD optimizer is set to 0.1 . I check the parameter value after updating:  

-0.2994 - 0.1 * (-1.6246) =  -0.1369, 

the result is as expected.

### DIFF
--- a/torch/torch.go
+++ b/torch/torch.go
@@ -6,9 +6,16 @@ package torch
 // #include "cgotorch.h"
 import "C"
 import (
-	"github.com/wangkuiyi/gotorch/aten"
+	"reflect"
 	"unsafe"
+
+	"github.com/wangkuiyi/gotorch/aten"
 )
+
+// Optimizer struct
+type Optimizer struct {
+	Opt C.Optimizer
+}
 
 // RandN returns a tensor filled with random number
 func RandN(rows, cols int, requireGrad bool) aten.Tensor {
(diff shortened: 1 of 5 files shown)


torch/torch.go
```diff
@@ -6,10 +6,17 @@ package torch
...
 import (
-	"github.com/wangkuiyi/gotorch/aten"
+	"reflect"
 	"unsafe"
+
+	"github.com/wangkuiyi/gotorch/aten"
 )
 
+// Optimizer struct
+type Optimizer struct {
```
wangkuiyi, line 16: It seems that we can move the newly added content into a new file `optimizer.go`.

torch/torch.go
```diff
@@ -18,3 +25,33 @@ func RandN(rows, cols int, requireGrad bool) aten.Tensor {
...
+
+// ZeroGrad reset gradients to zero
+func (opt Optimizer) ZeroGrad() {
+	C.ZeroGrad(opt.Opt)
+}
+
+// Step updates parameters
+func (opt Optimizer) Step() {
+	C.Step(opt.Opt)
+}
```
wangkuiyi, line 57: It seems that we need `Optimizier.Close()` to delete the underlying C++ optimizer?

QiJune: Yes, I will fix it in next PR.

torch/torch.go
QiJune: Yes, there will be other types of optimizers, like adam/adagrad/etc. we could move them into `optimizer.go`.

PR #36, opened 2020-07-31
use Setfinalizer to manage memory

QiJune:


examples/sgd/sgd.go
```diff
@@ -18,11 +20,8 @@ func main() {
 		loss.Backward()
 		opt.Step()
 
-		loss.Close()
-		pre.Close()
-		b.Close()
+		runtime.GC()
+		runtime.GC()
```
QiJune, line 24: Here, we must call `runtime.GC()` twice.

Here is the log of calling once:

```log
0
1
Close Tensor, 10, 100
Close Tensor, 10, 100
2
Close Tensor, 10, 100
3
Close Tensor, 10, 100
4
Close Tensor, 10, 100
5
6
Close Tensor, 10, 100
Close Tensor, 10, 100
7
Close Tensor, 10, 100
8
Close Tensor, 10, 100
9
Close Tensor, 10, 100
Close Tensor, 100, 10
```

Here is the log of calling twice:

```log
0
Close Tensor, 10, 100
1
Close Tensor, 10, 100
2
Close Tensor, 10, 100
3
Close Tensor, 10, 100
4
Close Tensor, 10, 100
5
Close Tensor, 10, 100
6
Close Tensor, 10, 100
7
Close Tensor, 10, 100
8
Close Tensor, 10, 100
9
Close Tensor, 10, 100
Close Tensor, 100, 10
```

torch.go
```diff
@@ -16,27 +17,36 @@ type Optimizer struct {
...
 
 // RandN returns a tensor filled with random number
-func RandN(rows, cols int, requireGrad bool) Tensor {
+func RandN(rows, cols int, requireGrad bool) *Tensor {
 	rg := 0
 	if requireGrad {
 		rg = 1
 	}
-	return Tensor{C.RandN(C.int(rows), C.int(cols), C.int(rg))}
+	t := &Tensor{C.RandN(C.int(rows), C.int(cols), C.int(rg))}
```
QiJune, line 25: We must return `*Tensor` here to make sure that the tensor is created at heap. 
Otherwise, calling `runtime.GC` will destroy all the objects. The training loop would be panic.

examples/sgd/sgd.go
```diff
@@ -1,15 +1,17 @@
...
+	"runtime"
+
 	torch "github.com/wangkuiyi/gotorch"
 )
 
 func main() {
 	a := torch.RandN(100, 10, true)
 	opt := torch.NewSGDOpt(0.1, 0, 0, 0, false)
-	opt.AddParameters([]torch.Tensor{a})
+	opt.AddParameters([]*torch.Tensor{a})
```
wangkuiyi, line 12: Why do we need * here?

examples/sgd/sgd.go
wangkuiyi: Good catch. It seems that we can keep once in the program but keep it in our mind that double-call clears cleanly.

tensor.go
```diff
@@ -5,46 +5,55 @@ package gotorch
...
+import "runtime"
 
 // Tensor wrappers a pointer to C.Tensor
 type Tensor struct {
 	T C.Tensor
 }
 
 // String returns the Tensor as a string
-func (a Tensor) String() string {
+func (a *Tensor) String() string {
```
wangkuiyi, line 16: Why do we need to change to use * here?

torch.go
```diff
@@ -46,16 +56,16 @@ func (opt Optimizer) AddParameters(tensors []Tensor) {
 }
 
 // ZeroGrad reset gradients to zero
-func (opt Optimizer) ZeroGrad() {
+func (opt *Optimizer) ZeroGrad() {
 	C.ZeroGrad(opt.Opt)
 }
 
 // Step updates parameters
```
wangkuiyi, line 63: Why do we need* here?

torch.go
wangkuiyi: I don't quite understand this point. What does"all the objects" refer to?

QiJune: @wangkuiyi 

Following is the original `RandN`

```
func RandN(rows, cols int, requireGrad bool) Tensor {
	rg := 0
	if requireGrad {
		rg = 1
	}
	t := Tensor{C.RandN(C.int(rows), C.int(cols), C.int(rg))}
	runtime.SetFinalizer(&t, func(f *Tensor) {
		f.Close()
                fmt.Printf("Close tensor, %d, %d\n", rows, cols)
	})
	return t
}
```
It's returned by value.



Then, in following training loop:

```
a := torch.RandN(100, 10, true)
opt := torch.NewSGDOpt(0.1, 0, 0, 0, false)
opt.AddParameters([]*torch.Tensor{a})

for i := 0; i < 10; i++ {
        fmt.Println(i)
	b := torch.RandN(10, 100, false)
	pre := torch.MM(b, a)
        runtime.GC()
}
```

I call `runtime.GC()`, and expect to free the memory of tensor b only. But it will free the memory of tensor a and tensor b both. The error log is following:

```log
0
Closed tensor 10, 100
Closed tensor 100, 10
1
libc++abi.dylib: terminating with uncaught exception of type c10::Error: Expected a Tensor of type Variable but found an undefined Tensor for argument #1 'mat2' (checked_cast_variable at ../torch/csrc/autograd/VariableTypeManual.cpp:38)
frame #0: c10::Error::Error(c10::SourceLocation, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 135 (0x4291f47 in libc10.dylib)
frame #1: torch::autograd::VariableType::unpack(at::Tensor const&, char const*, int) + 374 (0x11228a96 in libtorch_cpu.dylib)
frame #2: torch::autograd::VariableType::mm(at::Tensor const&, at::Tensor const&) + 937 (0x10c60ed9 in libtorch_cpu.dylib)
frame #3: c10::detail::wrap_kernel_functor_unboxed_<c10::detail::WrapRuntimeKernelFunctor_<at::Tensor (*)(at::Tensor const&, at::Tensor const&), at::Tensor, c10::guts::typelist::typelist<at::Tensor const&, at::Tensor const&> >, at::Tensor (at::Tensor const&, at::Tensor const&)>::call(c10::OperatorKernel*, at::Tensor const&, at::Tensor const&) + 21 (0xeb23615 in libtorch_cpu.dylib)
frame #4: at::Tensor c10::KernelFunction::callUnboxed<at::Tensor, at::Tensor const&, at::Tensor const&>(c10::OperatorHandle const&, at::Tensor const&, at::Tensor const&) const + 165 (0x422c5b5 in libcgotorch.so)
frame #5: at::Tensor c10::Dispatcher::callUnboxedWithDispatchKey<at::Tensor, at::Tensor const&, at::Tensor const&>(c10::OperatorHandle const&, c10::DispatchKey, at::Tensor const&, at::Tensor const&) const + 171 (0x422c2db in libcgotorch.so)
frame #6: at::Tensor c10::Dispatcher::callUnboxed<at::Tensor, at::Tensor const&, at::Tensor const&>(c10::OperatorHandle const&, at::Tensor const&, at::Tensor const&) const + 158 (0x422c19e in libcgotorch.so)
frame #7: at::Tensor c10::OperatorHandle::callUnboxed<at::Tensor, at::Tensor const&, at::Tensor const&>(at::Tensor const&, at::Tensor const&) const + 94 (0x422c0ee in libcgotorch.so)
frame #8: at::mm(at::Tensor const&, at::Tensor const&) + 154 (0x4217c9a in libcgotorch.so)
frame #9: MM + 33 (0x4217b91 in libcgotorch.so)
frame #10: _cgo_bdf0e44227b3_Cfunc_MM + 33 (0x40a25b1 in sgd)
frame #11: runtime.asmcgocall + 112 (0x405cec0 in sgd)
```

QiJune: Thanks to @shendiaomo 's help, let me clarify it more clear.

Here is original RandN code:
```go
func RandN(rows, cols int, requireGrad bool) Tensor {
	rg := 0
	if requireGrad {
		rg = 1
	}
	t := Tensor{C.RandN(C.int(rows), C.int(cols), C.int(rg))}
	runtime.SetFinalizer(&t, func(f *Tensor) {
		f.Close()
                fmt.Printf("Close tensor, %d, %d\n", rows, cols)
	})
	return t
}

a := RandN(100, 10, false)
```
`t` is created at the stack, after calling `runtime.GC`, `SetFinalizer` will be called, and the memory is released.
`a` has the same value as `t`. However, the memory is released already, `C.Tensor` of `a` become a dangling pointer.

I made a change:

```go
func RandN(rows, cols int, requireGrad bool) *Tensor {
	rg := 0
	if requireGrad {
		rg = 1
	}
	t := &Tensor{C.RandN(C.int(rows), C.int(cols), C.int(rg))}
	runtime.SetFinalizer(&t, func(f *Tensor) {
		f.Close()
                fmt.Printf("Close tensor, %d, %d\n", rows, cols)
	})
	return t
}

a := RandN(100, 10, false)
```

This time, `t` is created at heap. `a` is a pointer to the Tensor. When calling `runtime.GC`, the lifetime of `a` does not end. It will not be released. The behavior is expected.

tensor.go
```diff
@@ -5,46 +5,55 @@ package gotorch
...
+	runtime.SetFinalizer(t, func(f *Tensor) {
+		f.Close()
+	})
+	return t
 }
 
 // Sum returns the sum of all elements in the input tensor
-func Sum(a Tensor) Tensor {
-	return Tensor{C.Sum(a.T)}
+func Sum(a *Tensor) *Tensor {
```
shendiaomo, line 53: 改成这样是否能规避这个问题？
```golang
struct Tensor {
    T *C.Tensor
}

func MM(a, b Tensor) Tensor {
    ct := &C.MM(a.T, b.T)
    runtime. SetFinalizer(ct, func(ct *C.Tensor)) {
        C.Tensor_Close(*ct)
    }
    return Tensor{ct}
}
```
这样ct作为指针类型，应该会由GC接管，传递时不会立刻不可达。

tensor.go
QiJune: Since we return `*Tensor` with `RandN` and  `*Optimizer` with `NetSGDOpt`, it's better to keep consistent.

version 1:

```go
a := torch.RandN(100, 10, false)
(*a).String()
```

version 2:
```go
a := torch.RandN(100, 10, false)
a.String()
```

Version 2 is simpler.

tensor.go
QiJune: 有道理，我试一下

examples/sgd/sgd.go
QiJune: I'm not sure. 

Let's imagine a case, a training program runs in a GPU for many batches and occupies all the GPU memory.  But at some time, the GC does not release GPU memory in time, the training program may be crashed due to OOM.

examples/sgd/sgd.go
```diff
@@ -18,11 +20,7 @@ func main() {
 		loss.Backward()
 		opt.Step()
 
-		loss.Close()
-		pre.Close()
-		b.Close()
+		runtime.GC()
```
shendiaomo, line 23: runtime.GC()应该放在循环开头，放在末尾的话，`b`, `pre`, `loss`还没有变成不可达，第一次迭代不会回收任何东西，到第二次迭代结束，才会一起回收前两次迭代分配的张量。

PR #39, opened 2020-08-03
Use sync.WaitGroup to make sure all finalizers are done

shendiaomo:
This is a stricter version of https://github.com/wangkuiyi/gotorch/pull/36
Because `optimizer`s have a different lifetime with `tensor`s, they are not recycled by `torch.GC()`

### DIFF
--- a/examples/sgd/sgd.go
+++ b/examples/sgd/sgd.go
@@ -9,6 +9,7 @@ func main() {
 	opt := torch.NewSGDOpt(0.1, 0, 0, 0, false)
 	opt.AddParameters([]torch.Tensor{a})
 
+	torch.PrepareGC()
 	for i := 0; i < 100; i++ {
 		torch.GC()
 		b := torch.RandN(10, 100, false)
@@ -19,4 +20,5 @@ func main() {
 		loss.Backward()
 		opt.Step()
 	}
+	torch.FinishGC()
 }
--- a/examples/sgd/sgd.go
+++ b/examples/sgd/sgd.go
@@ -9,7 +9,6 @@ func main() {
 	opt := torch.NewSGDOpt(0.1, 0, 0, 0, false)
 	opt.AddParameters([]torch.Tensor{a})
 
-	torch.PrepareGC()
 	for i := 0; i < 100; i++ {
 		torch.GC()
 		b := torch.RandN(10, 100, false)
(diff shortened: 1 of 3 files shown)


tensor.go
```diff
@@ -6,45 +6,87 @@ package gotorch
...
+		go func() {
+			C.Tensor_Close(*t)
+			cTensorFinalizersWG.Done()
+		}()
+	})
+}
+
+// PrepareGC should be called right before a train/predict loop
+func PrepareGC() {
+	gcPrepared = true
```
wangkuiyi, line 33: Why do we need gcPrepared?

examples/sgd/sgd.go
```diff
@@ -9,20 +9,16 @@ func main() {
 	opt := torch.NewSGDOpt(0.1, 0, 0, 0, false)
 	opt.AddParameters([]torch.Tensor{a})
 
+	torch.PrepareGC()
```
wangkuiyi, line 12: torch.BeginTraining ?

examples/sgd/sgd.go
```diff
@@ -9,20 +9,16 @@ func main() {
...
 		opt.Step()
-
-		loss.Close()
-		pre.Close()
-		b.Close()
 	}
-
-	opt.Close()
-	a.Close()
+	torch.FinishGC()
```
wangkuiyi, line 23: torch.EndTraining?

tensor.go
```diff
@@ -6,45 +6,87 @@ package gotorch
 // #include "cgotorch.h"
 import "C"
 
+import (
+	"runtime"
+	"sync"
+)
+
+var (
+	cTensorFinalizersWG = &sync.WaitGroup{}
```
wangkuiyi, line 15: tensorFinalizersWG

examples/sgd/sgd.go
shendiaomo: It seems this function 's not necessary.

examples/sgd/sgd.go
shendiaomo: The function may be called by a predict loop, we need a more accurate name.

tensor.go
shendiaomo: Done.

tensor.go
shendiaomo: We have to use this to mark variables with different lifecycles, i. e. variables live in and out of the train loop.

PR #41, opened 2020-08-04
Remove unnecessary std::move

shendiaomo:
The copy constructor is adequate for this scenario.

### DIFF
--- a/cgotorch/cgotorch.cc
+++ b/cgotorch/cgotorch.cc
@@ -1,5 +1,7 @@
 #include "torch/script.h"
 #include "torch/torch.h"
+
+// FIXME(shendiaomo): including cgotorch.h before torch/torch.h will fail
 #include "cgotorch.h"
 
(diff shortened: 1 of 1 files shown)


cgotorch/cgotorch.cc
```diff
@@ -8,25 +8,25 @@
 Tensor RandN(int rows, int cols, int require_grad) {
   at::Tensor t = torch::randn({rows, cols},
                               at::TensorOptions().requires_grad(require_grad));
-  return new at::Tensor(std::move(t));
+  return new at::Tensor(t);
```
QiJune, line 11: There is only one data member in `at::Tensor`,  and it's a shared pointer. So, the copy constructor is enough?

shendiaomo: Yes, the copy-ctor and move-ctor are generated automatically, they're both viable and have little difference in performance.

PR #56, opened 2020-08-05
Edit C++ code to be compatible with cpplint

wangkuiyi:
This PR makes pre-commit to call cpplint, cppcheck, and clang-format.

On your macOS, please run the following commands to install these tools.

- `pip install cpplint`
- `brew install cppcheck`
- `brew install clang-format`

and update local pre-commit hooks.

```bash
pre-commit install
```


wangkuiyi:
I removed cppcheck and clang-format and left only cpplint. The installation of cppcheck and clang-format in Travis CI macOS VM causes brew upgrade, which takes too long time.

PR #60, opened 2020-08-05
add Adam optimizer

QiJune:
Adam optimizer is needed in DCGAN example.

Besides, this PR adds a method `ToSlice`, which copies data from libtorch to Go. This is useful when users debugging codes.


cgotorch/cgotorch.cc
```diff
@@ -46,6 +48,21 @@ void Tensor_Print(Tensor a) {
 
 void Tensor_Close(Tensor a) { delete static_cast<at::Tensor *>(a); }
 
+void Tensor_Copy(Tensor a, float *data, int num) {
```
wangkuiyi, line 51: Is the only purpose of introducing Tensor_Copy to implement Tensor.ToSlice() for Go unit test? If so, I am not sure if the ROI is too low.  I assume that Adam optimizer has been unit tested in libtorch's C++ code.

QiJune: In Python,  converting a Torch Tensor to a NumPy array and vice versa is a breeze. Please refer to the [doc](https://pytorch.org/tutorials/beginner/blitz/tensor_tutorial.html#numpy-bridge).

In GoTorch, we also need a method to convert a libtorch Tensor to a Go array and vice versa.

 In DCGAN model, we also need to get a value from a pytorch tensor:

```python
 netG.zero_grad()
label.fill_(real_label)  # fake labels are real for generator cost
output = netD(fake)
errG = criterion(output, label)
errG.backward()
D_G_z2 = output.mean().item()
optimizerG.step()
```

D_G_z2 is a python value, it's converted from a PyTorch tensor.

PR #64, opened 2020-08-05
expose conv2d function

QiJune:


tensor.go
```diff
@@ -111,3 +112,38 @@ func Sum(a Tensor) Tensor {
 	setTensorFinalizer(&t)
 	return Tensor{&t}
 }
+
+// Conv2d does 2d-convolution
+func Conv2d(input Tensor, weight Tensor, bias Tensor, stride []int,
+	padding []int, dilation []int, groups int) Tensor {
+	if bias.T == nil {
```
QiJune, line 119: It's a little long-winded here. I am thinking about how to simplify it.

PR #65, opened 2020-08-05
Support basic error handling

shendiaomo:
Now we'll `panic` if an exception happens. In this PR only `torch.MM` is processed this way. Later I'll rewrite other functions

### DIFF
--- a/cgotorch/Makefile
+++ b/cgotorch/Makefile
@@ -7,7 +7,7 @@ macos/libtorch : libtorch-macos-1.5.1.zip
 	unzip -qq -o $< -d macos
 
 libcgotorch.so : cgotorch.cc cgotorch.h macos/libtorch
-	rm -f libtorch
+	rm -rf libtorch/
 	ln -s macos/libtorch libtorch
 	clang++ -std=c++14 \
 	-I .. \
--- a/cgotorch/Makefile
+++ b/cgotorch/Makefile
@@ -7,7 +7,7 @@ macos/libtorch : libtorch-macos-1.5.1.zip
 	unzip -qq -o $< -d macos
 
 libcgotorch.so : cgotorch.cc cgotorch.h macos/libtorch
-	rm -rf libtorch/
+	rm -f libtorch
 	ln -s macos/libtorch libtorch
 	clang++ -std=c++14 \
 	-I .. \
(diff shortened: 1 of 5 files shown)


cgotorch/cgotorch.h
```diff
@@ -9,8 +9,12 @@ extern "C" {
 
 typedef void *Tensor;
 typedef void *Optimizer;
+typedef struct {
+  Tensor t;
+  char *err;
+} Result;
 Tensor RandN(int rows, int cols, int require_grad);
-Tensor MM(Tensor a, Tensor b);
+Result MM(Tensor a, Tensor b);
```
wangkuiyi, line 17: I think the C-conventional way to return more than one values is to have an additional parameter, other than defining `Result` type.

```c
Tensor MM(Tensor a, Tensor b, char** err);
```

shendiaomo: Done.

cgotorch/cgotorch.cc
```diff
@@ -16,10 +16,18 @@ Tensor RandN(int rows, int cols, int require_grad) {
...
-  return new at::Tensor(c);
+char *MM(Tensor a, Tensor b, Tensor *result) {
+  try {
+    at::Tensor c =
+        at::mm(*static_cast<at::Tensor *>(a), *static_cast<at::Tensor *>(b));
+    *result = new at::Tensor(c);
+    return nullptr;
+  } catch (const std::exception& e) {
+    auto len = strlen(e.what());
+    auto r = new char[len + 1];
```
wangkuiyi, line 27: So every time we call MM, we need to call FreeString() to release the error right?  I don't see the call to FreeString in the examples?

cgotorch/Makefile
```diff
@@ -7,7 +7,7 @@ macos/libtorch : libtorch-macos-1.5.1.zip
 	unzip -qq -o $< -d macos
 
 libcgotorch.so : cgotorch.cc cgotorch.h macos/libtorch
-	rm -f libtorch
+	rm -rf libtorch/
```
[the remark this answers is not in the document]
wangkuiyi: I fixed it.

cgotorch/cgotorch.h
```diff
@@ -9,8 +9,12 @@ extern "C" {
 
 typedef void *Tensor;
 typedef void *Optimizer;
+typedef struct {
+  Tensor t;
+  char *err;
+} Result;
```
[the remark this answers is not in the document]
wangkuiyi: I fixed it.

PR #74, opened 2020-08-06
support more tensor init function

QiJune:
This PR also change `int` in C++ header/source file to `int64_t`.

### DIFF
--- a/cgotorch/cgotorch.cc
+++ b/cgotorch/cgotorch.cc
@@ -3,6 +3,7 @@
 #include <sstream>
 #include <string>
 #include <tuple>
+#include <unordered_map>
 #include <vector>
 
 #include "torch/script.h"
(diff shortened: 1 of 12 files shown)


cgotorch/cgotorch.cc
```diff
@@ -22,14 +23,83 @@ char *exception_str(const std::exception &e) {
   return r;
 }
 
+std::unordered_map<std::string, torch::nn::init::FanModeType> fan_mode_map = {
+    {"fan_in", torch::kFanIn},
+    {"fan_out", torch::kFanOut},
+};
+
+std::unordered_map<std::string, torch::nn::init::NonlinearityType>
```
wangkuiyi, line 31: It seems that this global variable is only used by `KaimingUniform_`. Should we move it into `KaimingUniform_` as a local static variable?

cgotorch/cgotorch.cc
```diff
@@ -22,14 +23,83 @@ char *exception_str(const std::exception &e) {
   return r;
 }
 
+std::unordered_map<std::string, torch::nn::init::FanModeType> fan_mode_map = {
```
wangkuiyi, line 26: This global variable is only used by `KaimingUniform_` as `non_linearity_map`.

cgotorch/cgotorch.h
```diff
@@ -10,7 +10,20 @@ extern "C" {
 
 typedef void *Tensor;
 typedef void *Optimizer;
-char *RandN(int rows, int cols, int require_grad, Tensor *result);
+// torch.randn
+char *RandN(int64_t *size, int64_t length, int64_t require_grad,
```
wangkuiyi, line 14: Should we make the return type from `char *` to `const char *`?

cgotorch/cgotorch.cc
QiJune: There is another parameter initialization named [`kaiming_normal_`](https://pytorch.org/cppdocs/api/function_namespacetorch_1_1nn_1_1init_1ac8a913c051976a3f41f20df7d6126e57.html).  It also uses these two maps. So let's take it as global variables.

cgotorch/cgotorch.h
QiJune: Yes, we do not modify the return error message. `const char*` is more accurate.

PR #85, opened 2020-08-07
Use forward declaration to leverage the C++ type system

shendiaomo:
This PR removes most of the `static_cast`s to ensure type safety.


wangkuiyi:
Thanks for this simplification!!

PR #87, opened 2020-08-09
Make Travis CI use ARM64 other than AMD64

wangkuiyi:


wangkuiyi:
Travis CI doesn't support ARM, but supports ARM64. However, https://github.com/ljk53/pytorch-rpi provides the ARM version libtorch but not ARM64.

PR #92, opened 2020-08-10
add nn.functional directory

### LINKED ISSUE #90: mkdir nn/functional and move modules.go there

QiJune:
Fix #90 

This PR also adds [`nn.initializer`](https://pytorch.org/docs/stable/nn.init.html) directory.


nn/functional/functional.go
```diff
@@ -0,0 +1,102 @@
...
+			cweight,
+			cbias,
+			cmean,
+			cvar,
+			cTraining,
+			C.double(momentum),
+			C.double(eps),
+			cCudnnEnabled,
+			&t)))
+	torch.SetTensorFinalizer((*unsafe.Pointer)(&t))
```
wangkuiyi, line 51: Should we make SetTensorFinalizer a method of Tenosr, say Tenosr.SetFinalizer?

QiJune: The `SetTensorFinalizer ` method takes a `*C.Tensor` as input.

If we make a `Tensor. SetFinalizer ` method, we have to write codes as follows:

```go
var t C.Tensor
MustNil(unsafe.Pointer(C.RandN((*C.int64_t)(unsafe.Pointer(&shape[0])),
	C.int64_t(len(shape)), C.int64_t(rg), &t)))
tt := Tensor{(*unsafe.Pointer)(&t)}
tt.SetFinalizer()
return tt
```

The codes have 5 lines, which is longer than the current implementation(4 lines).

```go
var t C.Tensor
MustNil(unsafe.Pointer(C.RandN((*C.int64_t)(unsafe.Pointer(&shape[0])),
	C.int64_t(len(shape)), C.int64_t(rg), &t)))
SetTensorFinalizer((*unsafe.Pointer)(&t))
```

PR #95, opened 2020-08-10
Define the  Module interface and the `Sequential` and `Functional` containers

shendiaomo:
1. This PR has both `IModule` and `Module`, the former is the interface, the latter is to ease implementation.
    - The `IModule` interface exposes much fewer methods than in the C++/Python `Module` class, other methods will be added later when necessary.
2. This PR adds the `Sequential` container, whose `Forward` method is based on `reflect`. We use `reflect` rather than polymorphism to implement the `Forward` method to give users complete freedom to design their `Forward` methods. This behavior is the same as the Python/C++ frontend of PyTorch.
    - Therefore, the `getNamedNonNilTensors` function in module.go is also modified to support introspecting the `slice`s in an `IModule`.
    - This PR also adds a `Sequential` based example of MNIST for demo.

### DIFF
--- a/nn/module.go
+++ b/nn/module.go
@@ -7,9 +7,80 @@ import (
 	torch "github.com/wangkuiyi/gotorch"
 )
 
-// Module interface
-type Module interface {
-	Forward(x torch.Tensor) torch.Tensor
+// Dtype is the data type of scalars
+type Dtype int // TODO(shendiaomo): this is a placeholder to be defined later
+
+// DeviceType is the type of devices
+type DeviceType in // TODO(shendiaomo): this is a placeholder to be defined latert
+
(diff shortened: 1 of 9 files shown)


nn/module.go
```diff
@@ -1,38 +1,174 @@
...
+// Module contains default implementation of `Module`s
+type Module struct {
+	outer IModule
+	// Whether the module is in training mode.
+	isTraining bool
+	// The module's name (e.g. "LSTM").
+	name string
+}
+
+// Init initializes a `Module`, using a `Module` that's not `Init`ed is undefined behavior
```
wangkuiyi, line 49: We need to be able to check if a module is not initialized, and warns users about it. We cannot accept "undefined behavior".

PR #105, opened 2020-08-11
Add a simple MNIST benchmark

wangkuiyi:
The GoTorch version has a higher throughput that is 2~3 times higher than the PyTorch version.  Also, the GoTorch version has much shorter total running time than the PyTorch one.

On my quad-core iMac, we can see that the GoTorch version takes significantly higher ratio of system time than the PyTorch version.  The second peak in the following figure corresponds to the run of the GoTroch version; the third one to the PyTorch one.

2041597175957_ pic

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -21,8 +21,8 @@ of the GoTorch program is about **3 times** of the PyTorch version.  The measure
 of throughput doesn't include model instantiaation, data preparation, but only
 the train loop.
 
-Consider the tatal running time, the GoTorch version takes 43 seconds, but the
-PyTorch version takes 4 minutes.
+Consider the tatal running time, the GoTorch version takes **43 seconds**, but
+the PyTorch version takes about **4 minutes**.
 
 A typical run outputs the following:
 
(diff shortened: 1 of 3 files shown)


wangkuiyi:
Following @bddppq 's suggestion and set the following MKL and OMP thread numbers.

```
export MKL_NUM_THREADS=1
export OMP_NUM_THREADS=1
``` 

The performance of GoTorch and PyTorch versions get much closer to each other.

```
yi@WangYis-iMac:~/go/src/github.com/wangkuiyi/gotorch (benchmark)$ time go test -run TrainMNIST
2020/08/11 15:45:56 Throughput: 5576.162111 samples/sec
PASS
ok  	github.com/wangkuiyi/gotorch	54.985s

real	0m58.789s
user	0m54.431s
sys	0m3.071s
yi@WangYis-iMac:~/go/src/github.com/wangkuiyi/gotorch (benchmark)$ time python benchmark/mnist.py
The throughput: 4302.422927288301 samples/sec

real	1m11.339s
user	1m8.675s
sys	0m1.719s
```

PR #110, opened 2020-08-12
port torch.mean

QiJune:


tensor.go
```diff
@@ -168,6 +168,14 @@ func (a Tensor) Item() float32 {
 	return t
 }
 
+// Mean torch.mean
+func (a Tensor) Mean() Tensor {
```
wangkuiyi, line 172: Do we also need to follow the libtorch convention to have an additional method that doesn't create and return a new Tensor?

```go
func (a *Tensor) Mean_() {
  mutate a but don't create a new tensor
}
```

PR #111, opened 2020-08-12
Update benchmark report

wangkuiyi:


benchmark/README.md
```diff
@@ -57,3 +57,50 @@ ok  	github.com/wangkuiyi/gotorch	33.026s
...
+they have similar throughput -- in particular, the GoTorch version is 22\%
+faster than the PyTorch version.
+
+**Why is the C++ version runs that much faster than the Go version?**
+
+The reasons include:
+
+1. A C++ function call usually takes a few ns; however, it takes about 15μm for
+   a Go function to call a C/C++ function.
+1. The C++ program frees tensors out-of-scope using reference count maintained
```
shendiaomo, line 103: This is probably not the reason. the `GC`-caused overhead include:
1. Mark-and-sweep the unreachable `Go` variables
    - In a `gotorch` program, there're only a few `Go` variables.
2. Call C++ `new` upon cgo function call and `delete` upon the end of an iteration.
    - Only the `delete`s and `new`s are additional overhead in each iteration.

It seems that GC does not cause much overhead

wangkuiyi: Agreed. We should dig in for more plausible reasons,.

PR #120, opened 2020-08-13
DCGAN example

QiJune:
C++ DCGAN model reference: https://github.com/pytorch/examples/blob/master/cpp/dcgan/dcgan.cpp

Compile and Run:

```bash
cd cgotorch && make && cd ..

cd example 
go run dcgan.go
```

The fake image tensors will be saved into pickle files. Then we use a python script to convert the pickle files to images.

```bash
python to_image.py
```


example/dcgan.go
```diff
@@ -0,0 +1,222 @@
...
+	"github.com/wangkuiyi/gotorch/nn/initializer"
+)
+
+const (
+	images = "train-images-idx3-ubyte"
+	labels = "train-labels-idx1-ubyte"
+	site   = "http://yann.lecun.com/exdb/mnist/"
+)
+
+func downloadMNIST() error {
```
wangkuiyi, line 25: The source code for downloading MNIST data duplicates with that in mnist_test.go. Should we extract the code into a new package `gotorch/mnist` to reuse it?

QiJune: Done. Add a new package `gotorch/vision`. 
In Python, there is a torchvision package, it contains all vision dataset related util functions.

example/to_image.py
```diff
@@ -0,0 +1,15 @@
+import torch
```
wangkuiyi, line 1: pickle_image_to_png.py

PR #134, opened 2020-08-17
support cuda 101

QiJune:
Some Linux machines have CUDA 10.1 installed.


cgotorch/build.sh
```diff
@@ -10,15 +10,21 @@ if [[ "$OS" == "linux" ]]; then
         echo "Building for Raspbian ...";
         make -f Makefile.rpi;
     elif $(whereis cuda | cut -f 2 -d ' ')/bin/nvcc --version > /dev/null; then
-        echo "Building for Linux with CUDA ...";
-        make -f Makefile.linux-gpu;
+	CUDA_VERSION=`nvcc --version | grep release | grep -Eo "[0-9]+.[0-9]+" | head -1`
+	if [[ "$CUDA_VERSION" == "10.1" ]]; then
```
wangkuiyi, line 14: I introduced this script file in the hope that we can reduce the number of makefiles. In my mind, this script can download the corresponding libtorch zip file and set an environment variable to the name of the downloaded zip file, which can then be used by the only one makefiles to unzip.

Also, since makefiles.rpi requires gcc other than clang, we might want to unify to use gcc.

QiJune: Good suggestion! I put downloading command into the bash scripts, and pass variables into Makefile.

PR #140, opened 2020-08-17
Create design.md

wangkuiyi:


doc/design.md
```diff
@@ -0,0 +1,316 @@
...
+   defining the forward computation process and data members that can store
+   states.
+
+## Tensors and Garbage Collection
+
+libtorch includes the C++ definition of the fundamental data type `at::Tensor`
+and native functions that operate it.
+
+The key design feature of the tensor is automatic garbage collection (GC). In
+C++, the class `at::Tensor` contains only one data member, a `shared_ptr` to the
```
shendiaomo, line 48: It's indeed a `c10::intrusive_ptr`, which is an intrusive version of `shared_ptr`, with better performance.

doc/design.md
```diff
@@ -0,0 +1,316 @@
...
+cyclic-dependency.
+
+PyTorch programmers access `at::Tensor` from the Python binding. Python’s GC
+uses reference counts like `shared_ptr`.  For completeness, Python runs
+mark-and-sweep from time to time to handle cyclic-dependencies.
+
+Go provides an asynchronous API, `runtime.GC()`, to trigger GC and returns
+immediately without waiting for the completion of GC.  If all tensors are in CPU
+memory, this mechanism works; however, in deep learning, we would prefer to host
+tensors in GPU memory, which is a precious resource. We prefer free tensors
```
QiJune, line 62: prefer free --> prefer to free

doc/design.md
```diff
@@ -0,0 +1,316 @@
...
+`To` and `ZeroGrad`, which must know what tensors need to be moved to the
+specified device, or which tensors are parameters and whose gradient tensors
+need to be cleared.
+
+Similarly, GoTorch allows users to define modules in Go. Users define a struct
+type, say `LinearModule` that embeds the struct `gotorch/nn.Module`, which has
+useful utility methods attached.
+
+We don’t want to bother users to call the `register_xxx` function to specify the
+role of eac data member.  So, in the definition of type `LinearModule`, we allow
```
QiJune, line 242: eac --> each

doc/design.md
```diff
@@ -0,0 +1,316 @@
...
+libtorch includes the C++ definition of the fundamental data type `at::Tensor`
+and native functions that operate it.
+
+The key design feature of the tensor is automatic garbage collection (GC). In
+C++, the class `at::Tensor` contains only one data member, a `shared_ptr` to the
+implementation of the tensor type.  The smart pointer `shared_ptr` implements
+reference count-based GC, which frees a tensor once its reference count becomes
+zero.  Comparing with Go and Java’s GC, which runs the mark-and-sweep algorithm,
+reference count reacts instantly but cannot handle the case of
+cyclic-dependency.
```
shendiaomo, line 53: `std::weak_ptr` can resolve cyclic-dependency of `std::shared_ptr`s with careful design. `c10::weak_intrusive_ptr` is the intrusive version of `std::weak_ptr`.

doc/design.md
```diff
@@ -0,0 +1,316 @@
...
+
+Users can define PyTorch modules in Python and C++, but mostly in Python.
+
+To define a new module in Python, users derive a class from `torch.nn.Module`,
+adding data members corresponding to parameters, buffers, and sub-modules, and
+attaching the method `forward` that defines the forward pass of the module.
+
+The Python module-definition framework have Python decorators that call C++ core
+functions `Module::register_{parameter|buffer|module}`.  This registration
+mechanism is necessary as later users might call the module class’ methods like
```
QiJune, line 232: The Python module-definition framework and C++ module-definition framework are completely independent. Python module-definition framework does not call C++ `register_parameter
|buffer` functions.

In C++, users have to call `Module::register_parameter|buffer|module` explicitly. Following is an example of defining a Linear module.

```cpp
void LinearImpl::reset() {
  weight = register_parameter("weight",
    torch::empty({options.out_features(), options.in_features()}));
  if (options.bias()) {
    bias = register_parameter("bias", torch::empty(options.out_features()));
  } else {
    bias = register_parameter("bias", {}, /*requires_grad=*/false);
  }
  reset_parameters();
}
```

In Python, [`register_parameter`](https://github.com/pytorch/pytorch/blob/v1.6.0/torch/nn/modules/module.py#L289) method is hide behind [`__setattr__`](https://github.com/pytorch/pytorch/blob/v1.6.0/torch/nn/modules/module.py#L774) method. So, when adding a data member to the module class, the `register_parameter` in Python will be called implicitly.

```python
def __init__(self, in_features: int, out_features: int, bias: bool = True) -> None:
        super(Linear, self).__init__()
        self.in_features = in_features
        self.out_features = out_features
        self.weight = Parameter(torch.Tensor(out_features, in_features))
        if bias:
            self.bias = Parameter(torch.Tensor(out_features))
        else:
            self.register_parameter('bias', None)
        self.reset_parameters()
```

doc/design.md
```diff
@@ -0,0 +1,316 @@
...
+    return nullptr;
+  } catch (const std::exception &e) {
+    return exception_str(e.what());
+  }
+}
+```
+
+We see that `torch::randn` defined in libtorch returns an `at::Tensor`. However,
+the Cgo wrapper `RandN` returns a C string representing the possible C++
+exception.  The last parameter of `RanN` is the returned tensor.
```
shendiaomo, line 149: `RanN` -> `RandN`

doc/design.md
wangkuiyi: Thank you for the very comprehensive examples. I am restructuring the section, embedding them into the new version.

doc/design.md
```diff
@@ -0,0 +1,352 @@
...
+## Tensors and Garbage Collection
+
+libtorch includes the C++ definition of the fundamental data type `at::Tensor`
+and native functions that operate it.
+
+The key design feature of the tensor is automatic garbage collection (GC). In
+C++, the class `at::Tensor` contains only one data member, a smart pointer to
+the real tensor object.  This smart pointer performs reference count-based GC,
+which frees a tensor once its reference count gets zero.  Comparing with Go and
+Java’s GC, which runs the mark-and-sweep algorithm, reference count reacts
```
byzhang, line 52: Compared to Go and Java’s GC which run the mark-and-sweep algorithm, reference count reacts ...

doc/design.md
```diff
@@ -0,0 +1,352 @@
...
+instantly but cannot handle the case of cyclic-dependency.
```
byzhang, line 53: Is cyclic-dependency an issue for smart pointers?

doc/design.md
```diff
@@ -0,0 +1,352 @@
...
+   backward pass in each step of the train loop.
+
+The customized Go GC mechanism doesn’t handle the first two categories, which is
+the topic of the next section, Porting Modules.
+
+To handle intermediate results, GoTorch users need to call `gotorch.GC()` at the
+beginning of each train loop step.  The first job of `gotorch.GC()` is to mark
+that all tensors generated since then, which are considered intermediate
+results, are subject to the customized GC.  After the train loop, users are
+supposed to call `gotorch.FinishGC()` to unset the mark.
```
byzhang, line 91: why do users need to call `FinishGC`? All masked tensors should have been recycled when exiting loops?

doc/design.md
wangkuiyi: Done

doc/design.md
wangkuiyi: It is a problem with strong reference-based smart pointers like shared_ptr.  Users need to carefully choose to use `week_ptr` somewhere in the program to handle cyclic-dependencies.

doc/design.md
wangkuiyi: The function `torch.GC` is supposed to be invoked at the beginning of an iteration, as it enables the marking of subsequently created tensors to be intermediate results.

It might be more comprehensive to write

```go
torch.BeginGC()
for ... {
    // the train loop
    torch.GC()
}
torch.EndGC()
```

QiJune (approved):
Excellent!

PR #146, opened 2020-08-18
Add CircleCI

wangkuiyi:
CircleCI runs way faster than Travis CI because CircleCI runs in Docker containers instead of VMs. However, the benefit of running VM is to add macOS and ARM64 tests.

Should we have CircleCI runs Linux without CUDA tests, and, optionally, Travis CI runs macOS and ARM tests?


QiJune (approved):
Excellent！

PR #151, opened 2020-08-19
Use pprof for benchmark analysis

wangkuiyi:
1. On my iMac, the GoTorch version of MNIST training has throughput close to that of the C++ counterpart.
1. On my iMac, the GoTorch and C++ versions are times faster than the PyTorch version.  As previously noticed and analyzed, the is likely due to that pip-installed PyTorch doesn't have a highly optimized libtorch shipped with it.
1. The GoTorch version spends 70% of runtime in libtorch and 21% on Cgo call -- Both are under expectation.

I removed code that calls `runtime.trace` added in https://github.com/wangkuiyi/gotorch/pull/138, because what we want is not tracing but pprof.

There is no need to change the test code to have pprof; instead, we can run the following command to get the profiling output into `cpu.prof`.

```bash
go test -cpuprofile cpu.prof -memprofile mem.prof -v -run TrainMLPUsingMNIST
```

Then, we can run `go tool pprof -pdf cpu.prof` to render `gpu.prof` into a PDF file.

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -26,7 +26,7 @@ type the following command:
 make -C benchmark && /usr/bin/time ./benchmark/mnist
 ```
 
-All the above programs use only CPU but no GPU.  We will add GPU support in the 
+All the above programs use only CPU but no GPU.  We will add GPU support in the
 next
 week.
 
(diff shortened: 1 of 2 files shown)


Yancey1989 (approved):
LGTM

PR #158, opened 2020-08-20
Add a simple training loop of resnet50

QiJune:
**Naive version**

Set batch size to 16

GoTorch GPU memory: 4016MiB
PyTorch GPU memory: 2555MiB

PyTorch uses `nn.ReLU(inplace=True)`, while GoTorch not. This needs to be optimized further.

**After adding Relu inplace**

GoTorch GPU memory: 3326MiB

**After adding Add inplace**

GoTorch GPU memory: 2942MiB

### DIFF
--- a/cgotorch/torch.cc
+++ b/cgotorch/torch.cc
@@ -137,6 +137,15 @@ const char *Add(Tensor a, Tensor other, float alpha, Tensor *result) {
   }
 }
 
+const char *Add_(Tensor a, Tensor other, Tensor *result) {
+  try {
+    *result = new at::Tensor(a->add_(*other));
+    return nullptr;
+  } catch (const std::exception &e) {
+    return exception_str(e.what());
+  }
+}
+
 const char *Flatten(Tensor a, int64_t startDim, int64_t endDim,
                     Tensor *result) {
   try {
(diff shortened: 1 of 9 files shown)


tensor.go
```diff
@@ -320,6 +320,17 @@ func Add(a, other Tensor, alpha float32) Tensor {
 	return Tensor{(*unsafe.Pointer)(&t)}
 }
 
+// AddInplace torch.add_
+func (a *Tensor) AddInplace(other Tensor) Tensor {
```
wangkuiyi, line 324: How about we also name it `Add_`?

cgotorch/torch.cc
```diff
@@ -137,6 +137,15 @@ const char *Add(Tensor a, Tensor other, float alpha, Tensor *result) {
   }
 }
 
+const char *Add_(Tensor a, Tensor other, Tensor *result) {
+  try {
+    *result = new at::Tensor(a->add_(*other));
```
wangkuiyi, line 142: To what I understand, here `a->add_` calls `aten::Tensor::add_`, which returns the reference to `a`, but not a newly created tensor. Am I correct?

If so, it seems that we don't need to call `SetTensorFinalizer` in the Go method `Tensor.AddInplace`?

https://github.com/wangkuiyi/gotorch/pull/158/files#diff-ae7d2f2494c2de4f64de14717ee14fbdR330

PR #159, opened 2020-08-20
Rename Tensor.AddInplace into Tensor.AddI

wangkuiyi:
Address https://github.com/wangkuiyi/gotorch/pull/158/files#r474089268

In libtorch, there are three Add function/methods:

1. `Tensor& Tensor::Add_(Tensor)` returns the tensor itself.
1. `Tensor Tensor::Add(Tenosr)`
1. `Tensor Add(Tensor, Tensor)`

In Go, we can have

1. `func (t Tensor) AddInplace(Tensor)`
1. `func (t Tensor) Add(Tensor) Tensor`
1. `func Add(a, b Tensor) Tensor`

The only question is `AddInplace` is too long, and `Add_` is prohibitive by Go linter. How about one of the following?

- `Add∅` -- ∅ represents an empty set.
- `AddØ` -- Ø is a Norwegian letter considered the origin of ∅.
- `Add0` -- 0 is easy to type but might be misleading -- adding 0 to the current tensor?

### DIFF
--- a/tensor.go
+++ b/tensor.go
@@ -320,8 +320,8 @@ func Add(a, other Tensor, alpha float32) Tensor {
 	return Tensor{(*unsafe.Pointer)(&t)}
 }
 
-// אdd torch.add_
-func (a *Tensor) אdd(other Tensor) Tensor {
+// Add adds in-place
+func (a *Tensor) Add(other Tensor) Tensor {
 	var t C.Tensor
 	MustNil(unsafe.Pointer(C.Add_(
 		C.Tensor(*a.T),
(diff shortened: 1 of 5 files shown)


tensor.go
```diff
@@ -320,8 +320,8 @@ func Add(a, other Tensor, alpha float32) Tensor {
 	return Tensor{(*unsafe.Pointer)(&t)}
 }
 
-// AddInplace torch.add_
-func (a *Tensor) AddInplace(other Tensor) Tensor {
+// Add adds in-place
+func (a *Tensor) Add(other Tensor) Tensor {
```
QiJune, line 324: According to the discussion, we use `AddI` to represent the Add inplace.

Then we will have:

```
func (t Tensor) AddI(Tensor) Tensor
func (t Tensor) Add(Tensor) Tensor
func Add(a, b Tensor) Tensor
```

PR #160, opened 2020-08-20
Add documents on wrapping PyTorch native functions

wangkuiyi:


doc/wrap_native_functions.md
```diff
@@ -0,0 +1,156 @@
...
+    return nullptr;
+  } catch (const std::exception &e) {
+    return exception_str(e.what());
+  }
+}
+```
+
+This C++ implementation is the one called by the Go version of MM under the name
+`C.MM`.  You might wonder why not Go MM calls the native function `mm` directly
+via `C.mm`, but `C.MM` in `cgotorch/torch.cc`.  The only reason is we need to
```
ljk53, line 140: The native functions have C++ declarations. My understanding is that we need provide C wrappers around C++ functions for go to call. So to answer 'why not Go MM calls the native function directly', besides exception handling, C++ objects need to be wrapped as C pointers for golang to manage their destruction explicitly. So we manually dereference C.Tensor to at::Tensor before calling c++ functions and manually copy the result into a heap tensor before returning.

doc/wrap_native_functions.md
```diff
@@ -0,0 +1,209 @@
...
+The C wrapper `MM` returns a string serialization of the possible C++ exception
+thrown by libtorch if there is any, or `nullptr`.
+
+The implementation of the C wrapper `MM` is in C++ and in `cgotorch/torch.cc`.
+
+```cpp
+const char *MM(Tensor a, Tensor b, Tensor *result) {
+  try {
+    at::Tensor c = at::mm(*a, *b);
+    *result = new at::Tensor(c);
```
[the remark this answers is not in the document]
wangkuiyi: No. It doesn't, because here the new operator calls the [move constructor](https://en.cppreference.com/w/cpp/language/move_constructor) instead of the copy constructor. Let me explain this in the doc.

doc/wrap_native_functions.md
```diff
@@ -0,0 +1,213 @@
...
+}
+```
+
+At first glance, this definition looks too much more complicated than the C type
+Tensor, which is a pointer.
+
+The literal translation of the C type Tensor into Go is
+
+```go
+type Tenosr unsafe.Pointer
```
QiJune, line 171: Tenosr --> Tensor

doc/wrap_native_functions.md
```diff
@@ -0,0 +1,213 @@
...
+
+1. It passes the address of `t` of type `C.Tensor` to the C wrapper, `C.MM`.
+   Because `C.Tensor` is a pointer to `at::Tensor`, the line `*result = new
+   at::Tensor(c)` in the C wrapper makes `t` pointing to the newly allocated
+   tensor on the heap.
+1. It calls `MustNil` to check the C string returned from the C wrapper.
+   `MustNil` panics if the string is not `nullptr`.
+1. It calls `gotorch.SetTensorFinalizer` to attach the finalizer to the tensor
+   returned by `C.MM`.
+1. It returns a value of the Go type `Tenosr` that encapsulates `t`.
```
QiJune, line 213: Tenosr --> Tensor

doc/wrap_native_functions.md
```diff
@@ -0,0 +1,213 @@
...
+```go
+// #include <stdlib.h>
+import "C"
+func MyExit(x int) {
+    C.exit(x)
+}
+```
+
+We put all C wrappers of native functions in the subdirectory `cgotorch`.  In
+`cgotorch/cgotorch.h`, we can see the wrapper of `at::Tenosr` and `at::mm`.
```
QiJune, line 111: at::Tenosr --> at::Tensor

doc/wrap_native_functions.md
```diff
@@ -0,0 +1,213 @@
...
+
+```c
+extern "C" {
+typedef at::Tensor* Tensor;
+const char *MM(Tensor a, Tensor b, Tensor *result);
+}
+```
+
+C does not have classes, but C has pointers, so we define C type `Tenosr` as a
```
QiJune, line 120: C type `Tenosr` --> C type `Tensor`

doc/wrap_native_functions.md
```diff
@@ -0,0 +1,213 @@
...
+   doesn't copy the content of `c` because the operator `new` calls the
+   [move](https://en.cppreference.com/w/cpp/language/move_constructor)
+   constructor instead of the copy constructor of `at::Tensor`.
+1. It returns the string-serialization of the exception if there is any, or
+   `nullptr`.
+
+## Go Wrappers and Go Tensor
+
+In package `gotorch`, we define the Go wrappers of native functions, which
+operates the Go type `Tenosr` defined in `tensor.go`.
```
QiJune, line 156: Go type `Tenosr` --> Go type `Tensor`

QiJune (approved):
Excellent!

doc/wrap_native_functions.md
```diff
@@ -0,0 +1,213 @@
...
+    return exception_str(e.what());
+  }
+}
+```
+
+It runs the following steps, which are in most other wrappers too.
+
+1. It calls the native function `mm` and creates the result tensor `c` on the
+   stack.
+1. It allocates a heap object `*result` and **moves** `c` to `*result`.  This
```
shendiaomo, line 145: I believe the code snippet above calls *copy-ctor* rather than *move-ctor* because the variable `at::Tensor c` would match an *lvalue const reference* in the signature of the *copy-ctor*: https://github.com/pytorch/pytorch/blob/029007c8b650c84267ac1cdaae3bba8e5a9f957b/aten/src/ATen/templates/TensorBody.h#L92

Nevertheless, the result is the same in both cases: `c` is destructed and `*result` holds the only reference to the underlying `TensorImpl` object upon returning.

wangkuiyi: I see. I should explain that the destruction and re-construction is very efficient as `at::Tensor` contains only a smart pointer -- am I correct?

shendiaomo (approved):

PR #194, opened 2020-08-25
Add more unit tests

wangkuiyi:
- Split tensor.go into smaller files.
- Add unit tests using NewTensor and by comparing it with Python program outputs.
- Numerical result comparison using Tensor.String.


QiJune (approved):
Excellent!

PR #217, opened 2020-08-26
Add cifar10 dataset

QiJune:
The mean and std of normalization transform are float64, as a result, the tensor after normalization will be cast to float64 too. It's not expected.

The training data and parameter are usually float32 type.


vision/datasets/cifar10.go
```diff
@@ -0,0 +1,109 @@
...
+	"image/color"
+	"io/ioutil"
+	"os"
+	"path"
+
+	torch "github.com/wangkuiyi/gotorch"
+	"github.com/wangkuiyi/gotorch/vision/transforms"
+)
+
+var cifar10Url = "https://www.cs.toronto.edu/~kriz/cifar-10-binary.tar.gz"
```
wangkuiyi, line 15: It would be great if we can have a comment here explaining the format of that tar.gz file.

vision/transforms/to_tensor.go
```diff
@@ -39,8 +39,14 @@ func imageToTensor(img image.Image) torch.Tensor {
 		for y := 0; y < width; y++ {
 			// ResNet need the 3 channels image, here we should convert to RGB format.
 			// The division by 255.0 is applied to convert RGB pixel values from [0, 255] to [0.0, 1.0] range
-			c := img.(*image.NRGBA).NRGBAAt(x, y)
-			row[y] = [3]float32{float32(c.R) / 255.0, float32(c.G) / 255.0, float32(c.B) / 255.0}
+			switch img := img.(type) {
```
wangkuiyi, line 42: We don't need these type assertions. We should use the interface type `image.Image` and `image.Color` instead.  Please refer to https://github.com/wangkuiyi/gotorch/pull/222 for how I removed such type assertions.

QiJune: Done, remove it.

vision/datasets/cifar10.go
QiJune: Done, add a link here.

PR #222, opened 2020-08-26
Fix to tensor bug

wangkuiyi:
- Remove the type assertion without checking of success or failure.
- `torch.Int` means Go's `int32`, not `int`.
- It always fails to convert a Go slice-of-slice into a tensor using FromBlob.

### DIFF
--- a/vision/transforms/to_tensor.go
+++ b/vision/transforms/to_tensor.go
@@ -37,10 +37,8 @@ func imageToTensor(img image.Image) torch.Tensor {
 	for x := 0; x < height; x++ {
 		row := make([][3]float32, width)
 		for y := 0; y < width; y++ {
-			// ResNet need the 3 channels image, here we should convert to RGB format.
-			// The division by 255.0 is applied to convert RGB pixel values from [0, 255] to [0.0, 1.0] range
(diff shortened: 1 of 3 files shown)


vision/transforms/to_tensor.go
```diff
@@ -30,27 +30,29 @@ func (t ToTensorTransformer) Run(obj interface{}) torch.Tensor {
...
-
-	for x := 0; x < height; x++ {
-		row := make([][3]float32, width)
-		for y := 0; y < width; y++ {
-			// ResNet need the 3 channels image, here we should convert to RGB format.
-			// The division by 255.0 is applied to convert RGB pixel values from [0, 255] to [0.0, 1.0] range
-			c := img.(*image.NRGBA).NRGBAAt(x, y)
-			row[y] = [3]float32{float32(c.R) / 255.0, float32(c.G) / 255.0, float32(c.B) / 255.0}
+	array := make([]float32, h*w*3)
+	denom := float32(0xffff)
```
typhoonzero, line 36: Should be `float32(0xff)`, `float32(0xffff)` is 65535.

wangkuiyi: It is 0xffff. Not 0xff

PR #233, opened 2020-08-27
Add variation of Item()

lhw362950217:
We can get the value of a  0-dim tensor by `Item()`, however the data type in the tensor may be different. It can be float32, float64, int or many others.
We may return the value by an interface like: `func Item() interface{}`. However we still need to do dispatch in Go. So, I add some variation of the `Item` like `ItemInt64`, I think explicit type functions can make it easier in later usage.

### DIFF
--- a/cgotorch/tensor.cc
+++ b/cgotorch/tensor.cc
@@ -28,6 +28,24 @@ const char *Item(Tensor a, float *result) {
   }
 }
 
+const char *ItemInt64(Tensor a, int64_t *result) {
+  try {
+    *result = a->item<int64_t>();
+    return nullptr;
+  } catch (const std::exception &e) {
+    return exception_str(e.what());
+  }
+}
+
+const char *ItemFloat64(Tensor a, double *result) {
+  try {
+    *result = a->item<double>();
+    return nullptr;
+  } catch (const std::exception &e) {
+    return exception_str(e.what());
+  }
+}
+
 const char *Mean(Tensor a, Tensor *result) {
   try {
     *result = new at::Tensor(a->mean());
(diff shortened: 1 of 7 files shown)


cgotorch/tensor.cc
```diff
@@ -28,6 +28,24 @@ const char *Item(Tensor a, float *result) {
...
+const char *ItemInt64(Tensor a, int64_t *result) {
+  try {
+    *result = a->item<int64_t>();
+    return nullptr;
+  } catch (const std::exception &e) {
+    return exception_str(e.what());
+  }
+}
+
+const char *ItemFloat64(Tensor a, double *result) {
```
wangkuiyi, line 40: If we are going to have ItemInt64, ItemInt32, ItemInt16, ItemUint64, ItemUint32, ItemUint16 ... , there would be too large the interface.

I would prefer learning from https://golang.org/pkg/reflect/#Value to have

-  `func (v Value) Float() float64` to return floating point values. Users can down-cast to float32.
- `func (v Value) Int() int64` to return interger values.  Users can down-cast to Int32.
- `func (v Value) Uint() uint64` to return unsigned integer values.

lhw362950217: Agreed, so how about define the api as:
```go
func (t Tensor) Item() Value {
    if t.DType is int(64|32|16) {
      return wrapAsIntValue(C.ItemInt64(t))  // give 64bit integer to hold the value in C, we still need one C function each for int/uint/float type
    } else if t.DType is float(32|64) {
      return wrapAsFloatValue(C.ItemFloat64(t))
    } ...
}
```

wangkuiyi: The idiomatic way in Go to return the value of an arbitrary type is to represent the type by `interface{}` other than `reflect.Value`.

```go
package main

import (
	"fmt"
)

func f(choice int) interface{} {
	switch choice {
	case 1:
		return float32(0.9)
	case 2:
		return float64(0.8)
	case 3:
		return int64(7)
	case 4:
		return int16(6)
	}
	return nil
}

func main() {
	fmt.Println(f(3).(int64))
}
```

PR #234, opened 2020-08-28
port normal initializer

QiJune:


nn/initializer/initializer_test.go
```diff
@@ -14,3 +14,9 @@ func TestManualSeed(t *testing.T) {
 	expected := float32(0.66135216)
 	a.Equal(expected, x.Item())
 }
+
+func TestNormal(t *testing.T) {
+	x := torch.Empty([]int64{2, 3}, false)
+	Normal(&x, 0.1, 0.2)
+	assert.NotNil(t, x.T)
```
wangkuiyi, line 21: We need to compare specified values in unit tests, not simply checking no panics.

QiJune: It's not very straightforward to verify a random generation function. One way is to statistic the output and verify if the output value obeys the normal distribution. I will update a detailed unit test in the next PRs.

PR #235, opened 2020-08-28
Fix cgotorch/build.sh to use regexp match the arm arch

wangkuiyi:
- Use bash `=~` for regexp matching, because `$(uname -m)` prints `armvl7` instead of `arm` on Raspberry Pi.
- Use bash condition `[[ ]]` instead of that of sh `[ ]`.


QiJune (approved):
LGTM

PR #274, opened 2020-09-01
Train DCGAN with celeba dataset

QiJune:


example/dcgan/README.md
```diff
@@ -27,22 +31,34 @@ The `dcgan` binary will be installed at `$GOPATH/bin` directory.
...
+
+![example1](1.png) ![example2](2.png) ![example3](3.png)
+
+And the script could also generate an animation to visualize
+the training progress of generated fake images.
+`ffmpeg` is needed to save the animation to mp4 format.
 
 ```bash
-python pickle_to_png.py
+python visualize_pickle.py --save_video=1
```
wangkuiyi, line 63: Should we have a Go utility function to visualize a gotorch.Tensor into an image?

example/dcgan/dcgan.go
```diff
@@ -61,6 +63,45 @@ func discriminator(nc int64, ndf int64) *nn.SequentialModule {
 	)
 }
 
+func weightInit(m nn.IModule) {
+	if strings.Contains(m.Name(), "Conv") {
```
wangkuiyi, line 67: This looks a little hacky -- if a field name contains substring "Conv", then consider it an `nn.Module`.

nn/module.go
```diff
@@ -59,12 +63,53 @@ func (m *Module) Init(outer IModule) {
...
 	must(m.outer != nil, "GoTorch requires defining modules via embedding a `Module` struct by value")
 }
 
+// Name returns module name
+func (m *Module) Name() string {
+	return m.name
+}
+
+// Outer returns module outer
+func (m *Module) Outer() IModule {
```
wangkuiyi, line 79: The function `weightInit` looks a little bit hacky. I am not sure if it is worthy to change the interface `IModule` for it.

wangkuiyi:
BTW, we might not check in PNG files into Git repos.

example/dcgan/README.md
QiJune: I believe we should have.  And I think that it's not so urgent to add it right now. We could do this later.

example/dcgan/dcgan.go
QiJune: The Apply method will be added in the next PR.

nn/module.go
QiJune: The Apply method will be added in the next PR.

PR #291, opened 2020-09-03
Fix 0-dim shape calculation

lhw362950217:
Polish `Shape` function to support 0-dim tensor.


wangkuiyi (approved):
Thanks for this bug fixing!

Issue #288, opened 2020-09-02
Decoding jpg diff between Go image library and Python PIL library


QiJune:
I use the `ToTensor` transform to read the same image in GoTorch and PyTorch:

The last three Tensor value in GoTorch:
```
0.0788  0.0936  0.0936
```

In PyTorch:

```
0.0784, 0.0941, 0.0941
```

There is a little diff.

QiJune:
Please refer to #303

QiJune:
Here are two references: 

- https://stackoverflow.com/questions/44514897/difference-between-the-return-values-of-pil-image-open-and-tf-image-decode-jpeg
- https://stackoverflow.com/questions/44514897/difference-between-the-return-values-of-pil-image-open-and-tf-image-decode-jpeg

It seems that there are also differences between OpenCV, TensorFlow, and PIL.

> Well, this difference is arising from the fact that OpenCV, by default, uses integer accurate decompression of the JPEG image. In contrast, TensorFlow uses Discrete Cosine Transform as default.

wangkuiyi:
It seems that such differences are inevitable. Let us close the issue as we cannot do anything with it.

PR #297, opened 2020-09-05
Try to fix to_tensor precison

wangkuiyi:
Not sure if https://github.com/wangkuiyi/gotorch/issues/288 is due to the float precision problem -- it would be helpful if this issue describes how to reproduce the problem -- I guess this PR might be the reason.


QiJune:
@wangkuiyi Please refer to the #303 for reproducing the problem

wangkuiyi:
@QiJune Got it. Yeah. I verified float64/32 is not the reason for the difference. Close this PR as the issue is considered normal.

PR #313, opened 2020-09-08
Recycle memory in the To method

shendiaomo:


nn/module.go
```diff
@@ -123,6 +123,8 @@ func (m *Module) IsTraining() bool {
 // To recursively casts all parameters to the given `dtype` and `device`.
 func (m *Module) To(device torch.Device, dtype ...int8) {
 	must(m.outer != nil, "GoTorch requires calling `Init` before using")
+	torch.GC()
```
wangkuiyi, line 126: We need a comment explaining why we need this change.

shendiaomo: Done.

wangkuiyi (approved):

PR #319, opened 2020-09-09
Fix memory leak caused by Cgo thread spawning

### LINKED ISSUE #273: Memory leak while training Resnet50 model

shendiaomo:
Fix #273

### DIFF
--- a/tensor_gc.go
+++ b/tensor_gc.go
@@ -47,3 +47,10 @@ func GC() {
 	}
 	tensorFinalizersWG.Wait()
 }
+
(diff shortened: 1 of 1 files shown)


wangkuiyi (approved):
Because of the documentation of `runtime.LockOSThread` saying that we should call it before calling OS functionalities. I agree with this change.

Still, we need to figure out why adding this call fixes the many threads observed in a running GoTorch program.

PR #328, opened 2020-09-14
Reduce default number of threads to speed up training on GPUs

### LINKED ISSUE #321: runtime.LockOSThread() reduce throughput heavily
Training Resnet50 in P100.

Before #319,  the throughput of GoTorch is 70 samples/s. 

After #319, the throughput of GoTorch is 27 samples/s.

shendiaomo:
Fix #321 , test on P100 before merge.


device.go
```diff
@@ -28,3 +32,20 @@ func IsCUDAAvailable() bool {
...
 }
+
+// SetNumThreads sets the number of threads created by GoTorch
+func SetNumThreads(n int32) {
+	C.SetNumThreads(C.int32_t(n))
+}
+
+func init() {
+	// Avoid creating too many threads: the original default setting of OMP_NUM_THREADS
+	// (defaults to the core number) may degrade performance on GPUs.
```
wangkuiyi, line 43: Why OMP_NUM_THREADS = core number "may degrade" performance on GPUs?

Why should we change it to #core/2 ?

device.go
```diff
@@ -28,3 +32,20 @@ func IsCUDAAvailable() bool {
...
+	if os.Getenv("OMP_NUM_THREADS") == "" && os.Getenv("MKL_NUM_THREADS") == "" {
+		SetNumThreads(int32(runtime.NumCPU()) / 2)
+	}
+	// Prevent Cgo call from migrating to another system thread, hence the TLS cache in
+	// libtorch would not take too much RAM. See https://github.com/wangkuiyi/gotorch/issues/273
+	// for details.
+	runtime.LockOSThread()
```
wangkuiyi, line 50: Is the purpose here to prevent the goroutine scheduler from unparking too many threads when there the goroutine makes long-run Cgo calls?  If so, the comment needs to tell that 

> The goroutine scheduling has the following properties that might create too many OS threads.
> 1. The Go scheduler creates one or more P's, each represents a run queue, at program startup time.
> 1. The Go scheduler creates an OS thread M1 to run goroutines waiting in the run queue of a P.
> 1. If a goroutine X makes a Cgo call, and the call takes a long time (>20ms), the Go scheduler would be afraid that it will go on occupying that thread for more time, so it creates a new OS thread M2 for P to run other goroutines in the run queue.
> 1. After the Cgo call completes, the goroutine X goes back to the run queue of the P.  The next time it gets a turn to run, it runs on M2 instead of reusing M1, because M2 is now the default thread of P.
> 1. If X makes a new long-run Cgo call again, the Go scheduler would let X keeps using M2, and creates M3 for P to run other goroutines in its run queue.
> 1. The above process may create too many threads.  It is true that those out-of-use would be "parked", if they use some TLS, the TLS still takes memory.

Yancey1989 (approved):
LGTM

QiJune (approved):
LGTM

PR #338, opened 2020-09-17
Typo fixes

zhiqwang:
The link without `https://` in the doc will lead to https://github.com/wangkuiyi/gotorch/blob/develop/doc/github.com/pytorch/xla


wangkuiyi (approved):
Thanks for the correction!

PR #341, opened 2020-09-18
remove permute transform

QiJune:
We do not need this transform. PyTorch does not have such a permute transform too.


Yancey1989 (approved):
LGTM

PR #342, opened 2020-09-18
add opencv dependency in CI

QiJune:


wangkuiyi (approved):

.travis.yml
```diff
@@ -7,6 +7,16 @@ go:
   - 1.14
 os: osx
 
+cache:
+  directories:
+    - /usr/local
```
wangkuiyi, line 12: > GoTorch, in particular, gotorch/vision/transforms, uses OpenCV for image data augmentation. In Travis CI macOS VMs, we need to install OpenCV and its many dependencies. This would trigger Homebrew to update itself.  Thsi updating would take about 20 minutes. To shrink the time cost, we make Travis CI to cache the /usr/local directory, where Homebrew installs software packages.

wangkuiyi (commented):
Please add this comment.

.travis.yml
QiJune: Done, add the comment.

Yancey1989 (approved):
LGTM

PR #349, opened 2020-09-22
use ReadFull instead of Read

QiJune:
Fix #346 

```txt
Read reads data into p. It returns the number of bytes read into p. 
The bytes are taken from at most one Read on the underlying Reader, hence n may be less than len(p).
To read exactly len(p) bytes, use io.ReadFull(b, p). At EOF, the count will be zero and err will be io.EOF.
```


Yancey1989 (approved):
LGTM

shendiaomo (approved):

PR #353, opened 2020-09-25
Close `gocv.Mat`s that we forgot to close

shendiaomo:
`gocv.Mat`s must be released manually. They didn't choose `SetFinalizer`, see https://github.com/hybridgroup/gocv/issues/411 for details.


Yancey1989 (approved):
Excellent!

PR #359, opened 2020-10-09
Improve build time

shendiaomo:
1. Split the makefile target into multiple targets.
1. Use `make -j` to parallelize the building process.


wangkuiyi (approved):
LGTM++

PR #366, opened 2020-10-20
WIP: Support basic data parallel

shendiaomo:


nn/parallel/parallel.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+// DataParallel Evaluates module(input) in parallel across the given devices.
+// If `devices` is not supplied, the invocation is parallelized across all available CUDA devices.
+// If `outputDevice` is supplied, the final, combined tensor will be placed on this device. If not, it defaults to the first device in devices.
+// In detail, this method performs the following four distinct steps:
+//    1. Scatter the input to the given devices,
+//    2. Replicate (deep clone) the model on each device,
+//    3. Evaluate each module with its input on its device,
+//    4. Gather the outputs of each replica into a single output tensor, located on the `outputDevice`.
```
QiJune, line 32: There are two approaches for data parallelism in for multi-GPU training:

- Single-Process Multi-GPU
- Per Process Per GPU

PyTorch [DistributedDataParallel](https://github.com/pytorch/pytorch/blob/master/torch/nn/parallel/distributed.py#L461) has proved that Per Process Per GPU is more efficient.

> Single-Process Multi-GPU is not the recommended mode for "
                "DDP. In this mode, each DDP instance operates on multiple "
                "devices and creates multiple module replicas within one "
                "process. The overhead of scatter/gather and GIL contention "
                "in every forward pass can slow down training. "
                "Please consider using one DDP instance per device or per "
                "module replica by explicitly setting device_ids or "
                "CUDA_VISIBLE_DEVICES.

QiJune: So,  scatter--> parallel apply --> gather is not suggested. Instead, we launch a training process for each device. Each training process does dataloading/forward/backward/allreduce/update individually.

